PR #14, opened 2013-04-08
Fix tr1 headers

skhaz:
Added #ifdef _MSC_VER guards to include tr1 headers


nobu-k:
Thank you. The problem is fixed in the latest master branch.

PR #53, opened 2014-01-20
std::unordered_map and std::unordered_set added, tested.

xunzhang:
In my project, I am using std::unordered_map(C++11), so I add std::unordered_map/std::unordered_set.
Code tested with msgpack_unordered_map_set_test.cpp in test folder.


nobu-k:
Fixed in the master branch. The new major version will be released shortly.

PR #93, opened 2014-07-03
Update README.md

elliotwoods:
actually it's really nothing like JSON in terms of usage
but i can see where you're coming from :)

(p.s. this of course isn't entirely a serious pull request)


nobu-k:
haha, I like this one but you should ask it at https://github.com/msgpack/msgpack first since this repository is only one of the implementations following the original msgpack spec :)

PR #108, opened 2014-08-31
Add VERSION_MICRO macro definition

buzztaiki:
Add VERSION_MICRO macro definition for checking some trivial changes between micro versions.


nobu-k:
Merged ref #237

PR #182, opened 2015-01-03
General Repo Cleanup

haneefmubarak:
- rewrote portions of readme
 - cleaned up markdown in some files
 - added all contributors thus far to authors file
 - removed and renamed misc markdown files


nobu-k:
Thank you for the pull request. It looks nice.

As for AUTHORS file, I think it's easier to have a contributors section in README.md and put a link to https://github.com/msgpack/msgpack-c/graphs/contributors there to show the list of contributors rather than (a little manually) maintaining AUTHORS file. What do you think?

haneefmubarak:
The main reason I went for a file is that some projects will just include the entire repo (so markdown links just look funny) and that the contributors page on GitHub is really, really long (and it'll just keep getting longer).

nobu-k:
I see. msgpack-c has the same problem. It's derived from a msgpack repository, which had implementations of many languages in a single repository. So, some contributors aren't directly involved in msgpack-c.

Anyway. Because msgpack-c is now separated from the original mixed repository, I think showing the contributors page won't cause the problem. Then, I'd prefer to have a link in README.md if there's no problem other than what you pointed out.

haneefmubarak:
... I actually pulled all of the names from the `msgpack-c` repository contributors list @ https://github.com/msgpack/msgpack-c/graphs/contributors. There's >50 names, and that list will just keep growing. It's also fairly common to have a `[CONTRIBUTORS/CREDITS/AUTHORS]` file in OSS projects (the Linux repo tries to maintain one).

Also, what happens if sometime in the future, `msgpack-c` moves away from GitHub? Then the new list wherever might be missing some names while the GH list will no longer work.

I don't mind just adding a link to the README, but it just doesn't seem good for preserving credit for posterity.

nobu-k:
Thank you for sharing your concern.

I really like the contributors page of github in two ways. Firstly, it has more detailed information than regular contributors or authors file. For example, it shows when a contributor gave how much contribution to the repository. It's very good to see how they contributed, and at the same time it's difficult to manage such information manually. Secondly, it's automatically generated so that we won't forget to add a new great contributor to the list. We truly appriciate all contributions regard less of the size of patches, so we want to manage the list precisely. However, managing it manually often leads to some mistakes. Therefore, we want to rely on that feature. Although it's currently a little verbose and long, I believe the github team make it much more convenient in the future.

We can also create a AUTHORS file when it gets necessary. I think we'd likely use git command to generate a list of contributors as github does if we moved the repository to somewhere from github. So, at least for a while, I'd like to add a link to the README.

I'll merge this pull request later with an additional contributors page link in README.

PR #223, opened 2015-02-18
Allowing msgpack to build on FreeBSD

JackDanger:
FreeBSD doesn't have the byteswap header but __SVR4 is defined.


Pyrohh:
I think #222 already covers this.

JackDanger:
@Pyrohh Agreed. Closing.

PR #270, opened 2015-04-16
Inhibited wrong occurrence of ad-hoc encoding of 64bit doubles on iOS

makeroo:


redboltz:
@makeroo, thank you for sending the PR.
Let me clarify the situation. 

- __arm__ is defined and __ARM_EABI__ is NOT defined on iOS.
- However, iOS is NOT an oabi.
- Hence,  mem.i = (mem.i & 0xFFFFFFFFUL) << 32UL | (mem.i >> 32UL); is bad for iOS.
- To distinguish iOS, we can use TARGET_OS_IPHONE.

Is that correct? if it is, I will merge the PR soon.

makeroo:
On Mon, Apr 20, 2015 at 7:44 AM, Takatoshi Kondo <notifications@github.com>
wrote:

> @makeroo <https://github.com/makeroo>, thank you for sending the PR.
> Let me clarify the situation.
>
>    - *arm* is defined and *ARM_EABI* is NOT defined on iOS.
>
> correct


>
>    - However, iOS is NOT an oabi.
>
>  I'm not sure but I think so

>
>    - Hence, mem.i = (mem.i & 0xFFFFFFFFUL) << 32UL | (mem.i >> 32UL); is
>    bad for iOS.
>
> correct


>
>    - To distinguish iOS, we can use TARGET_OS_IPHONE.
>
> correct

>
>
> Is that correct? if it is, I will merge the PR soon.
>

thanks

PR #346, opened 2015-08-17
Make example build optional with cmake

aberaud:
When MsgPack is built as a part of a larger project it may save a lot of time to not build examples every time.

Also it makes the build more robust by compiling less code (potential errors or issues in examples won't affect the build).


nobu-k:
It was so helpful thanks! I just wanted to have this at the moment (to ignore compile errors in examples haha.

PR #421, opened 2016-02-14
Replace stdint.h header in check_container_size_decl.hpp

ibell:
``#include <stdint.h>`` is invalid in MSVC 9 2008 .  Using the system dependent header does work in my testing with MSVC9, and should be safe for other compilers I think.


redboltz:
@ibell, you are right. Thank you!

ibell:
Thank you very much!  I am now all fixed on the MSVC versions I need to
compile with.

PR #583, opened 2017-04-03
unpack_template: Include unpack.h for missing msgpack_unpack_return type

jamessan:
9a113bb0cae7c3b5e4a79ac6c5ede8a1ca5df889 added uses of
MSGPACK_UNPACK_{NOMEM,PARSE}_ERROR to unpack_template.h.  It didn't
include unpack.h though, which is where these values are defined.

src/unpack.c already includes unpack.h, so this doesn't affect
msgpack-c.  However, msgpack-perl also uses unpack_template.h and can't
build against cpp-2.1.0 due to this change.

### DIFF
--- a/include/msgpack/unpack_template.h
+++ b/include/msgpack/unpack_template.h
@@ -8,6 +8,8 @@
  *    http://www.boost.org/LICENSE_1_0.txt)
  */
 
+#include "msgpack/unpack.h"
+
 #ifndef msgpack_unpack_func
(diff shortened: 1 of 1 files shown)


jamessan:
Actually, I'll open a PR for msgpack-perl to include unpack.h directly instead of unpack_define.h.

jamessan:
Although, it would be useful if unpack.h included unpack_define.h, like pack.h does.

PR #674, opened 2018-04-25
Fix the fuzzer by setting limits for depth and ext.

derwolfe:
This is a short lived bug in the fuzzer implementation in which the limits were not set correctly. It adds limits for depth and ext as detailed in https://github.com/msgpack/msgpack-c/blob/c6c31dc5cd56ad0f8bafd0db82e013aa92cea5c3/include/msgpack/v1/unpack_decl.hpp#L94-L95.

Credit to OSS-Fuzz


derwolfe:
Please feel free to leave this PR sitting while you work on the #676

PR #721, opened 2018-08-13
Export an interface target for the C++ header-only library

dennisklein:
This improves integration with C++ projects that import msgpack via its CMake package. See [interface library](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#interface-libraries) for more details.


herbrechtsmeier:
The interface library isn't supported by CMake version 2.8.12.

dennisklein:
Good point, I will add a condition.

@herbrechtsmeier I also dont want to compete with your PR, which contains much more improvements. But because there was no activity and I wanted to see this particular feature... If you and @redboltz want to pick up the work on your PR again, I am happy to close this one.

PR #837, opened 2020-02-02
replace codecov with coveralls

ygj6:
- Replace codecov with coveralls

It works for my repo. But I'm not sure whether it works for the `msgpack` repo.
- Install dependencies in the folders outside the workspace

The reports generated by `coveralls` will include dependencies if the dependencies are installed inside the workspace . So I installed `gtest` to `/usr/local/` and `boost` to `~/boost/`.
Why not install `boost` to `/usr/local/`? The `actions/cache` has no permission to do it.
- Extract function (installing gtest) into a script (gtest.sh)

### DIFF
--- a/.github/depends/boost.sh
+++ b/.github/depends/boost.sh
@@ -10,8 +10,8 @@ EOL
 
 build_boost()
 {
-  BASE=`pwd`/..
-  ./b2 -j4 --toolset=$1 --prefix=${BASE}/usr --libdir="${BASE}/usr/$1/lib$2" --with-chrono --with-context --with-filesystem --with-system --with-timer address-model=$2 install
+  prefix=$HOME/boost
+  ./b2 -j4 --toolset=$1 --prefix=$prefix --libdir="$prefix/$1/lib$2" --with-chrono --with-context --with-filesystem --with-system --with-timer address-model=$2 install
 }
 
 bit="64"
--- a/.github/depends/gtest.sh
+++ b/.github/depends/gtest.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+usage()
+{
+  cat <<EOL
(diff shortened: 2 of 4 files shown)


ygj6:
The reports of `coveralls` is here:
https://coveralls.io/github/msgpack/msgpack-c

redboltz:
I created #838 to test on msgpack repository.

redboltz:
I think that the coverage data uploaded successfully.
See
https://coveralls.io/builds/28513840

But when I tried to see the file coverage, I got 500 error.
For example clieck TREE include/msgpack/v3/parse.hpp, then  I got error.
I guess that this is a way to check file level coverage.

Could you try this?

ygj6:
Ok, I will look at it soon.

ygj6:
I can open the file `include/msgpack/v3/parse.hpp`. 
Have you signed in to `coveralls` with your github account?

redboltz:
I didn't sign in. I tried sing in, then I got the result.
It is a little confusing that I can see the directory tree without sign in but I can't see the code without sign in.

I think that it is acceptable because someone who want to know the detail of coverage, they usually have github account.

redboltz:
15 hours ago, codecov started to implement tokenless uploading functionality.

See

https://github.com/codecov/codecov-action/issues/29#issuecomment-581540357
https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/15

@ygj6 , I respect your work for coveralls, but I'm more familiar with codecov. I'd like to wait a little more to decide which one should be used.

ygj6:
It's a good news. In fact, I prefer the user interface of `codecov`.
Do you think these two parts are useful? If yes, I will resubmit them after fixing the `codecov`
- Install dependencies in the folders outside the workspace
- Extract function (installing gtest) into a script (gtest.sh).

PR #1036, opened 2022-10-11
Specify that the project uses only C

dundargoc:
In rare cases, such as using the clang-cl compiler on Windows, the build assumes that the project is both C and C++ based and will give an error if the variable CMAKE_CXX_COMPILER isn't defined. Explicitly stating it's a C project prevents this situation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ IF ((CMAKE_VERSION VERSION_GREATER 3.1) OR
     CMAKE_POLICY(SET CMP0054 NEW)
 ENDIF ()
 
-PROJECT (msgpack)
+PROJECT (msgpack C)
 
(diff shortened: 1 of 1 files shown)


Fuzzier:
The project has a depedency on C++ compiler, since the `test` component depends on `gtest`, which is a C++ project.
Thus, if the project is set as pure-C, the `test` component could not be built.
`gtest` does not have a pure-C interface, so it would be hard to refacter the `test` component into pure-C.

Suggestion:
If the main `CMakeLists.txt` is specified as pure-C, then it shall not call `ADD_SUBDIRECTORY (test)`.
`test\CMakeLists.txt` shall be made a standalone project, and be built by a separate run of `cmake`.

dundargoc:
Hmm, the effort needed to fix this was much higher than I anticipated. I don't think the amount of change needed is worth the fix. Closing.

PR #1058, opened 2023-03-24
adaptor: better diagnostic for converter

timblechmann:
we only use the default `convert` implementation, if a `msgpack_unpack` method actually exists. if that's not the case we get an `undefined template` error.


timblechmann:
compare #1051

PR #1142, opened 2024-10-08
Fixed `find_package` command in README

uyha:
Since the file `msgpack-cxx-config.cmake` is generated during installation, `msgpack-cxx` should be used instead of only `msgpack`.


realysy:
Yes, indeed.

redboltz:
Apologies, I completely overlooked this pull request. Merging it now.

PR #1154, opened 2025-04-10
Work around MSVC C++20 modules error for nested msgpack::object::with_zone

scrossuk:
When packaging msgpack into a C++20 module (by #include of msgpack in the global module fragment and then `export using ...;` for relevant names), MSVC (VS 17.13.5) reports the following:

```
error C2504: 'msgpack::v2::object': base class undefined
```

This is apparently due to with_zone being a nested type of msgpack::object; in order to work around this problem, with_zone is replaced with an alias to a struct defined outside of msgpack::object.

### DIFF
--- a/include/msgpack/v1/object.hpp
+++ b/include/msgpack/v1/object.hpp
@@ -32,11 +32,11 @@ struct object_kv {
     msgpack::object val;
 };
 
-struct object::with_zone : msgpack::object {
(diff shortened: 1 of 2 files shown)


redboltz:
Thank you for sendint the PR. It looks good to me. CI coverage error is not related the PR, so I will merge it.

PR #1162, opened 2025-07-03
Include type_traits before use of std:: members

jwbee:
msgpack-cpp relies on its users to include, directly or indirectly, the <type_traits> header before including msgpack headers. If they don't, it can't compile, because std::remove_volatile will not be declared. This fixes the issue by including it from msgpack, which should be fine now that msgpack-cpp requires C++11.


redboltz:
Thank you for sending the PR.
Merged.
(CI error is fixed by #1163 )

PR #1177, opened 2026-06-09
Set INSTALL_INTERFACE to CMAKE_INSTALL_INCLUDEDIR

opoplawski:
While packaging msgpack-c for Fedora, I needed to install into /usr/include/msgpack-c to avoid conflicting with the msgpack C++ package.  But then packages that used it would not look in the correct location for the msgpack.h file.  This fixes that issue.


redboltz:
@opoplawski  Thank you for sending the PR. Merged.

PR #1187, opened 2026-09-18
ci: publish SBOMs for release archives

mcc0nnell:
## Summary

- generate SPDX 2.3 and CycloneDX 1.7 SBOMs for published `c-*` and `cpp-*` releases
- allow maintainers to backfill or dry-run an existing release tag with `workflow_dispatch`
- scan the exact uploaded release archive rather than the repository checkout
- verify the downloaded archive against GitHub's recorded release-asset SHA-256 before scanning
- attach both SBOMs to the same release without overwriting already-published SBOM evidence
- derive the distribution URL from the release asset and identify the source with a `pkg:github` PURL

This keeps SBOM generation outside the C and C++ build systems and does not change normal builds or release archives.

## Why

The current releases publish the source archive but no machine-readable SBOM alongside it. Generating from the published archive gives downstream software-assurance tooling an explicit, release-bound component identity without adding a CMake or package-manager dependency to msgpack itself.

The workflow fails closed if the downloaded archive does not match GitHub's release-asset digest, so the SBOM is bound to the bytes actually published on the release.

The workflow lives on `master` because GitHub release-event and manual-dispatch workflows are sourced from the repository default branch, while the released code continues to come from the `c-*` / `cpp-*` tag.
## Publication behavior

Published SBOM assets are treated as immutable evidence. A rerun leaves an existing `.spdx.json` or `.cdx.json` asset unchanged and uploads only missing assets; there is no `--clobber`.

The manual `tag` input lets maintainers exercise the workflow against an existing release such as `c-7.0.2` or `cpp-9.0.0` without cutting a new release.

The workflow downloads a pinned `sbom-utility` 0.19.2 archive, verifies its SHA-256, and schema-validates both generated SBOMs before publication.

## Scope of the SBOM

For the current MessagePack source archives, Syft produces a thin product-identity SBOM rather than a rich native dependency graph. In local validation of both current release tracks, the CycloneDX documents contained the root metadata component with `components: 0` and `dependencies: 0`.

That is still useful release evidence: it binds project/version, supplier, license, repository/tag PURL, distribution URL, and the exact release SHA-256. It should not be read as a claim that the workflow has discovered every possible C/C++ transitive dependency.
## Validation

Tested against the current release assets:

- `c-7.0.2` / `msgpack-c-7.0.2.tar.gz` — SHA-256 `6ae50f69612871aa01de76bec904165cd2a2fc30ff9f653f2f60a663c5c1a86c`
- `cpp-9.0.0` / `msgpack-cxx-9.0.0.tar.gz` — SHA-256 `303d3a7321aee65eb9450db8a6c973954e00af34b88ba0c0aca236bc50bfb8a9`

For both release tracks:

- GitHub's recorded release-asset digest matches the downloaded archive
- workflow passes `actionlint` 1.7.12
- SBOM generation tested with Syft 1.52.0
- resulting SPDX 2.3 and CycloneDX 1.7 documents validate with CycloneDX `sbom-utility` 0.19.2
- the verified SHA-256 is present on the root component/package
- the root PURL is `pkg:github/msgpack/msgpack-c@<release-tag>`
- the distribution URL comes from GitHub's release-asset API rather than a hardcoded upstream URL
- metadata enrichment preserves and deduplicates scanner-produced hashes, references, and licenses
- `gh release download` was tested without a checkout/local git repository

The Anchore action is pinned to the full commit for v0.24.0 and Syft is version-pinned to v1.52.0.

## CI note

AppVeyor is the repository's existing pull-request check and is independent of this release-event workflow. It fails on `master` before project code because that branch contains no Visual Studio project or solution; this workflow was validated independently as described above.


redboltz:
@mcc0nnell Thank you for sendint the PR. 
I merged it.

NOTE: I added appveyor settings (on appveyor's web ui) that start only if `c_master` and/or `cpp_master` is changed.

